PR #4, opened 2020-09-25
[WIP] SYCL XC Integrator

wavefunction91:
Reopen of https://github.com/wavefunction91/GauXC/pull/1 now that code has been made public and branches have been worked out


wavefunction91:
For some reason. `GAUXC_ENABLE_HOST=OFF` breaks the shared lib build when SYCL enabled on JLSE

wavefunction91:
Removing `assert`'s in GauXC and ExchCXX removes the shared linking problem. There might be a solution to this, but its not worth the effort for the time being

wavefunction91:
This is a stale PR. The code needs to remain public to accompany the paper results, but the SYCL implementation requires a complete rewrite

PR #10, opened 2020-09-29
Sycl cleanup

abagusetty:
1. minor changes to sycl kernel build-in APIs
2. support batched gemm


src/integrator/sycl/collocation_device.cpp
```diff
@@ -18,15 +18,23 @@ namespace sycl       {
                                  const T *pts_device, T *eval_device,
                                  cl::sycl::queue *queue) {
 
-        cl::sycl::range<2> threads(16, 16);
-        cl::sycl::range<2> blocks(util::div_ceil(npts, threads[0]),
-                                  util::div_ceil(nshells, threads[1]) );
+        cl::sycl::range<3> threads(16, 16, 1);
```
wavefunction91, line 21: Did you find out what the SYCL backend agnostic way of determining these things is? I'm in the process of removing "magic" numbers in the CUDA impl as well, would be good to get them sorted out here

src/integrator/sycl/collocation_device.cpp
```diff
@@ -18,15 +18,23 @@ namespace sycl       {
...
+                                  util::div_ceil(nshells, threads[1]),
+                                  1);
 
         GAUXC_SYCL_ERROR( queue->submit([&](cl::sycl::handler &cgh) {
                 auto global_range = blocks * threads;
 
-                cgh.parallel_for( cl::sycl::nd_range<2>(global_range, threads),
-                    [=](cl::sycl::nd_item<2> item_ct) {
+                cgh.parallel_for(
+                    cl::sycl::nd_range<3>(cl::sycl::range<3>(global_range.get(2),
```
wavefunction91, line 30: Why not just construct the `blocks` / `threads` ranges in the correct order to begin with?

Also, didn't we decide that `nd_range` was overkill here (since there's no reductions)? Would it be better to just use a regular range here and let SYCL determine how it wants to schedule it?

src/integrator/sycl/mklsycl_extensions.cpp
```diff
@@ -6,12 +6,19 @@ namespace GauXC {
...
 namespace blas  {
 
+    template <typename T>
+    void increment_kernel( const T* X, T* Y , cl::sycl::nd_item<3> item_ct) {
+        const auto tid = item_ct.get_group(2);
+        if( tid < 1 ) (*Y) += (*X);
+    }
     template <typename T> void increment(const T *X, T *Y, cl::sycl::queue *queue) {
         GAUXC_SYCL_ERROR( queue->submit([&](cl::sycl::handler &cgh) {
-           cgh.single_task([=] () {
```
wavefunction91, line 11: Does `single_task` not work here for some reason? That seems like a much better approach than a full blown kernel launch

src/integrator/sycl/mklsycl_extensions.cpp
```diff
@@ -61,16 +76,19 @@ namespace blas  {
     void hadamard_product(cl::sycl::queue *syclQue, int M, int N, const T *A, int LDA,
                           T *B, int LDB) {
 
-        cl::sycl::range<2> threads(16, 16);
-        cl::sycl::range<2> blocks(util::div_ceil(M, threads[0]),
-                                  util::div_ceil(N, threads[1]));
+        cl::sycl::range<3> threads(16, 16, 1);
```
wavefunction91, line 79: This should be 2D, no?

src/integrator/sycl/sycl_extensions.hpp
```diff
@@ -5,18 +5,18 @@ namespace GauXC {
 namespace sycl  {
 
 __inline__
-double warpReduceSum(double val, cl::sycl::nd_item<3>& item_ct) {
+double warpReduceSum(double val, cl::sycl::nd_item<3> item_ct) {
```
wavefunction91, line 8: Is `item_ct` better passed by reference or value? It's inconsistent throughout

wavefunction91 (changes requested):

src/integrator/sycl/mklsycl_extensions.cpp
abagusetty: `single_task` is the appropriate one. I missed this from the revert.

src/integrator/sycl/mklsycl_extensions.cpp
abagusetty: Addressed. I have lost this one too from reverts.

src/integrator/sycl/sycl_extensions.hpp
abagusetty: Addressed. In all the cases, the `nd_item` or `item` has been passed as reference.

src/integrator/sycl/collocation_device.cpp
abagusetty: Done. Similar to `cuda`'s work

src/integrator/sycl/collocation_device.cpp
abagusetty: done. `range` and `nd_range` have been appropriately placed where needed.

wavefunction91 (approved):

PR #15, opened 2020-10-02
Update gauxc-gau2grid.cmake

evaleev:
to avoid hardwiring built tree path.


wavefunction91 (approved):

wavefunction91:
This is fine, it looks like gau2grid has the proper linkage semantics to handle this on install

PR #18, opened 2020-10-07
[SYCL] Change memory access for collocation in SYCL integrator

wavefunction91:
WIP, need to update collocation UTs with transposed ordering


wavefunction91:
Breaks CI, but merging into SYCL development to minimize dev streams - will fix locally

PR #19, opened 2020-10-12
Adds optimization from proxy application

dmclark17:
Ported proxy app changes from my end to GauXC. I am marking as a draft because I did not include the downstream changes from the matrix transpose edit so it is not currently correct.


wavefunction91:
@dmclark17 I've rebased this locally, do you want me to push directly or have you review in a separate branch?

dmclark17:
I think pushing directly here would work

wavefunction91:
@dmclark17 Up-to-date. You can toggle your compact collocation kernel implementation by uncommenting 
https://github.com/wavefunction91/GauXC/pull/19/files#diff-c52215967dc1bb88bf7adcefa6c71b05a94b16c7af12c87ff3ea889112716b3cR4

and changing the analogous kernel launch. FWIW, this increases the register usage (`-gencode sm_70,compute_70 -O3`) from 64 -> 88

src/integrator/cuda/cuda_eval_denvars.cu
```diff
@@ -77,44 +82,62 @@ __global__ void eval_uvars_gga_kernel( size_t           ntasks,
...
-    const double* bf_y_col = dbasis_y_eval_device  + tid_y*nbf;
-    const double* bf_z_col = dbasis_z_eval_device  + tid_y*nbf;
-    const double* db_col   = den_basis_prod_device + tid_y*nbf;
-
-    den_reg = bf_col[ tid_x ]   * db_col[ tid_x ];
-    dx_reg  = bf_x_col[ tid_x ] * db_col[ tid_x ];
-    dy_reg  = bf_y_col[ tid_x ] * db_col[ tid_x ];
-    dz_reg  = bf_z_col[ tid_x ] * db_col[ tid_x ];
-
+  __shared__ double den_shared[4][GGA_KERNEL_SM_BLOCK_X][GGA_KERNEL_SM_BLOCK_Y+1];
```
wavefunction91, line 85: I think we can switch these over to `warp_size` and `max_warps_per_thread_block` from 
https://github.com/wavefunction91/GauXC/blob/master/src/integrator/cuda/cuda_device_properties.hpp#L7

src/integrator/cuda/cuda_eval_denvars.cu
```diff
@@ -77,44 +82,62 @@ __global__ void eval_uvars_gga_kernel( size_t           ntasks,
...
+  for ( int bid_x = blockIdx.x * blockDim.x; 
+        bid_x < nbf;
+        bid_x += blockDim.x * gridDim.x ) {
+    
+    for ( int bid_y = blockIdx.y * GGA_KERNEL_SM_BLOCK_Y; 
+          bid_y < npts;
+          bid_y += GGA_KERNEL_SM_BLOCK_Y * gridDim.y ) {
+        
+      for (int sm_y = threadIdx.y; sm_y < GGA_KERNEL_SM_BLOCK_Y; sm_y += blockDim.y) {
+        den_shared[0][threadIdx.x][sm_y] = 0.;
```
wavefunction91, line 96: Given that the length of these register dot products is relatively small ( O(1k) max, O(1) min ), do you think that it might be better to just restrict each grid point to a particular warp (i.e. don't break across thread blocks)? This would avoid shared memory and the `atomicAdd`'s at the end?

I haven't tested how this would effect the performance, thoughts?

dmclark17: That sounds like it is worth exploring—being able to accumulate directly in registers to avoid shared memory and atomics could be beneficial.

I was also curious about the performance of this kernel on Taxol; for the proxy application I recall seeing a larger performance gain using shared memory. Before any changes I have the `xc::eval_density_gga_kernel` kernel taking a total of 369.60ms, and with the shared memory optimization it takes 133.79ms.

However, I am seeing the `GauXC::integrator::cuda::eval_uvars_gga_kernel<double>` kernel take 348.95ms on the master branch and 315.35ms in the pull request branch. The kernels look the same to me, so I am wondering if my initial numbers were incorrect, or if I am confusing the kernel with a different one.

dmclark17:
For the `collocation_device_masked_combined_kernel_deriv1` kernel, I am also seeing that the grid-stride version is slower (135ms vs. 152ms for Taxol). I think we can stick with your implementation as the increased register usage is detrimental.

src/integrator/cuda/cuda_eval_denvars.cu
wavefunction91: It's possible this is a regression due to the use of CUB
https://github.com/wavefunction91/GauXC/blob/master/src/integrator/cuda/cuda_extensions.hpp#L6

If you uncomment that line, it will enable straight-up warp reductions. This was just left over from some testing, let me know if that changes anything (I didn't see anything egregious, but I didn't do a full profile with these changes)

If we get a significant improvement from just using the warp level intrinsics, that's probably a good sign that avoiding shared mem / atomics will be largely beneficial. We may also want to unroll that loop to reduce register pressure as well

wavefunction91:
I'm wondering if this is might be due to the fact that we're striding on multiple dimensions, it might be possible that a stride along `x` (i.e. the grid points) might be beneficial while stride along `y` (shells) is detrimental, or visa versa.

src/integrator/cuda/cuda_eval_denvars.cu
dmclark17: Whoops, I lost the changes to the kernel launch parameters in the transfer. I had followed your suggestion to do something like the following, but forgot to included it in the PR.

```cpp
blocks.x = std::min( 32, util::div_ceil( max_npts,    threads.x ));
blocks.y = std::min( 32, util::div_ceil( max_nshells, threads.y ));
```

Using the direct warp reductions had a minor performance impact, going from 138.11ms to 133.92ms in total time for the eval_uvars_gga_kernel for Taxol. Sorry about that—I will push the block size change soon.

src/integrator/cuda/cuda_eval_denvars.cu
```diff
@@ -187,9 +187,9 @@ void eval_uvars_gga_device( size_t           ntasks,
                             XCTaskDevice<T>* tasks_device,
                             cudaStream_t     stream ) {
 
-  dim3 threads(warp_size, max_warps_per_thread_block, 1);
-  dim3 blocks( util::div_ceil( max_nbf , threads.x ),
-               util::div_ceil( max_npts , threads.y ),
+  dim3 threads( warp_size, max_warps_per_thread_block / 2, 1 );
```
wavefunction91, line 190: Is there a reason this won't run on 1024? Is it a register issue?

dmclark17: It should run with a 32x32 thread-block, but I found 32x16 to be a little faster

wavefunction91:
@dmclark17  Is there anything else you'd like to include in this PR?

wavefunction91 (approved):

PR #21, opened 2020-10-13
Adds pack/inc optimization

dmclark17:
Optimizes the submatrix packing and incrementing kernels. I am marking this as a draft because it is hand-tuned to the Taxol problem size and V100 L2 cache size. 

There are 2 main parts of the optimization:

- Adding the submatrix indices to the submatrix cut vector—this allows cuts from the same task to be done concurrently. Additionally there is some loop unrolling to increase the number of load operations in flight, and inline PTX is used for stores to try to reduce pollution of L2 cache. 
- Block the large matrix such that it fits in the L2 cache. For the Taxol problem, the large matrix is about 1000 by 1000, which is roughly 8MBs. I believe this was causing a low L2 hit rate on a V100, which has about 6MB of L2 cache. To solve this, I am breaking the big matrix into four 512 by 512 chunks and using a separate kernel launch to process each one. To simplify the bookkeeping for the kernel, I enforce breaks in the cuts at intervals of 512. Each kernel call is passed the starting and stopping points within the submatrix cut vector it needs to process.

TODO:
- [x] The submatrix cut vector stores 2 pairs for each cut, but this could be optimized into 3 values as the lengths are the same. In fact, the end point of the small matrix is not used.
- [x] The second part of the optimization needs to be cleaned up, as the block size of the matrix should depend on the size of the L2 cache of the device
- [x] Right now, it is hard coded to break the problem into fourths and piggybacks on the submatrix cut vector to store the starting and stopping points for the kernel launch. This needs to be cleaned up so it can handle any number of blocks and so it uses a proper input variable to pass the starting and stopping points. 
- [x] Investigate the performance of multiple kernel launches when there are many small cuts. While I am measuring a speed-up on Taxol, this technique might not scale well when the large matrix breaks into many different chunks. If the tasks are sufficiently localized then this might lead to thread-blocks doing no work for some kernel launches, which might motivate a more intelligent way to launch the work.

My measurements showed that both of these optimizations had an effect on the kernels. Even though it is tuned to the Taxol problem size, it sped up the ubiquitin problem by about 2x for each kernel. 

I will work on the TODOs listed above, but I wanted to get your feedback on the approach.


wavefunction91 (commented):
Overall it looks good, I think this is a good approach. When you say 2x for Ubiquitin, do you mean 2x for the kernel or full TTS?

src/integrator/cuda/cuda_inc_potential.cu
```diff
@@ -7,16 +7,25 @@ namespace cuda       {
...
+#define WARP_Y 1
+#define UNROLL_FACTOR 4
+#define EFF_UNROLL 4
+#define CUT_X 8
+#define CUT_Y 4
 
-  if( batch_id < ntasks ) {
 
+template <typename T>
+__global__ __launch_bounds__(1024, 1)
```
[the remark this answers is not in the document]
dmclark17: For an unroll factor of 4, there is no register spilling. 

I mainly had it there to prevent launch errors, as the thread-block is large and the register usage is pretty high.

dmclark17:
Unfortunately, 2x for the individual kernels.

src/integrator/cuda/cuda_inc_potential.cu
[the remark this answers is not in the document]
wavefunction91: Ok, thanks for clarifying

wavefunction91:
As these are dominant for the big problems, this still has the potential to have a large impact

wavefunction91:
@dmclark17 I haven't dug into the cause of this yet, but it looks like your changes to https://github.com/wavefunction91/GauXC/pull/21/files#diff-298977ee40a20ec2b41b3c2c95681d68687ad52b2b4afbc97afba95c1388b4eaR60 cause a hang in the unit tests. My suspicion is that this is due to the fact that we only test small problems in the UTs, and the blocking factor of 512 yields something incorrect. I'll look more at it tomorrow

dmclark17:
Interesting, I will take a look as well. How would you like to proceed with changes to the code that is common to all the backends? Right now I am adding an additional pair for each cut and this could be modified to be a single triplet per cut. Do you think the changes will be propagated to the other backends, or should I made a CUDA specific version with the changes?

wavefunction91:
My guess is that something analogous will get migrated to the other
backends at some point (to the extent that it can at all), but I can handle
that as the need arises. I'll give it some thought over the next few days.
If it turns out that we need to sandbox this, we can do that before merge
once the functionality is fleshed out a bit more.

wavefunction91:
@dmclark17 I've merged `master` locally (with testing fixes https://github.com/wavefunction91/GauXC/pull/27), but that also pulls in the row-major memory access optimizations. I can't see any conflicts, but I wanted to confirm with you whether you'd like me to push directly here or once you've finished more of the tasks?

src/integrator/cuda/cuda_inc_potential.cu
```diff
@@ -7,16 +7,25 @@ namespace cuda       {
...
+void inc_by_submat_combined_kernel( size_t           ntasks,
+                                    XCTaskDevice<T>* device_tasks,
+                                    T*               A,
+                                    size_t           LDA, 
+				    const int block_y,
```
wavefunction91, line 25: I noticed that calling in calling this kernel, you make 4 calls (00,01,10,11). Would there be any benefit to just making this a non-type template parameter such that the calling sig is known at compile time? It's unclear to me to what extend `nvcc` exploits `constexpr` type knowledge, but traditional C++ compilers would likely be able to optimize around this knowledge (especially with `if constexpr`)

dmclark17:
I think pushing directly here would work—just to keep the branch up to date. I have made progress with some of the tasks and will try to push updates soon!

src/integrator/cuda/cuda_inc_potential.cu
dmclark17: This was hard coded to be 4 calls initially as `A` was only being broken into 4 parts. However, for larger problem sizes, I am breaking the matrix into many different chunks and have a kernel launch for each one.

I am finding that the optimal chunk size varies depending on the problem size and L2 cache of the device, with smaller chunks leading to better cache performance, but also increasing the number of kernel launches and overhead. I am trying to work out some heuristics to choose the optimal size for a given problem. For Taxol the matrix is O(1k) and for Ubiquitin it is O(10k); can we expect most problems to fall within this range, or could there be even larger problems of interest?

wavefunction91: I believe our stretch targets are O(50k) - O(100k), but I don't currently have examples to that effect. There are also other problems we have to address before we can go that large anyways (i.e. the assumption that the density/VXC can simultaneously fit in device memory will likely not hold). But if we can write up the methodology as to how to select these hyper-parameters, we can address how to handle this generally in the future

dmclark17:
I added a generalization to the blocking scheme for larger matrices as well as a few minor optimizations for the packing/inc kernels themselves. 

- To store information about the mapping from matrix blocks to cut indices, I introduced another array into the device task structure. This structure is generated by the `gen_compressed_submat_map` function which now returns a tuple of vectors. 
- I changed the `submat_map` structure to store deltas instead of end points which might be causing issues with the host backend.

src/integrator/cuda/cuda_pack_density.cu
```diff
@@ -7,54 +10,84 @@ namespace cuda       {
...
+#pragma unroll
+        for (int k = 0; k < UNROLL_FACTOR; k++) {
+          val[k] = ABig_begin[I + (J + k*WARP_Y)*LDA];
+          address[k] = ASmall_begin + I + (J + k*WARP_Y) * LDAS;
+        }
+#pragma unroll
+        for (int k = 0; k < UNROLL_FACTOR; k++) {
+	  // Suggest that the result be evicted first. CUDA11 provides
+          // a non-PTX way to do this so we can update if we require CUDA11
+          asm ("st.global.cs.f64 [%0], %1;" :: "l"(address[k]), "d"(val[k]));
```
dmclark17, line 79: In CUDA 11.0+ we can replace the inline-PTX with intrinsic functions.
https://docs.nvidia.com/cuda/archive/11.0/cuda-c-programming-guide/index.html#stx-functions

wavefunction91: I'll have to look into whether or not they have full support for CUDA 11 on Summit (they have a module, but iirc its experimental viz a few features). We can always just put this in a CPP check for CUDA version though (then deprecate once CUDA 11 is widely adopted @ HPC centers)

wavefunction91 (approved):

PR #30, opened 2020-10-29
Optimizes weights kernel

dmclark17:
This changes the level of parallelism to assign a warp to each point instead of a thread. 

* For the iParent portion, a warp computes the contribution from 32 jCenters concurrently. They then do a warp reduction to compute the final contribution. This changes the early exit scheme since the check effectively only happens every 32 jCenters.

* For the pairs portion, each thread in the warp computes the contribution from 1 iCenter at a time as the warp synchronously iterates over jCenters. The threads are able to independently exit early and start on a new iCenter.


dmclark17:
I updated the PR with some clean up changes and minor optimizations. 

Right now it is failing 371 of the assertions in the weight unit test, however the differences are all relatively small. Removing the reciprocal optimization seems to resolve the issue.

wavefunction91:
@dmclark17 How small is small? I'm more than happy to update the UT checks if its a worth while optimization

dmclark17:
The maximum absolute difference 1.4e-5 and the average absolute difference being 4.8e-8. The maximum percent difference is 0.4% and the average percent difference is 0.03%. 

The reciprocal optimization brings the total runtime of the weights kernel from 18.9s to 14.2s for a ubiquitin simulation on a V100.

src/integrator/cuda/cuda_weights.cu
```diff
@@ -14,33 +14,59 @@ namespace cuda       {
...
 
-  const int tid_x = threadIdx.x + blockIdx.x * blockDim.x;
-  const int tid_y = threadIdx.y + blockIdx.y * blockDim.y;
-
-
-  if( tid_y < natoms and tid_x < npts ) {
-
-    const int iAtom = tid_y;
-    const int iPt   = tid_x;
+  const int natoms_block = (natoms + 31) / 32;
```
wavefunction91, line 35: I would calculate these relative to the `warp_size` if possible

src/integrator/cuda/cuda_weights.cu
```diff
@@ -360,12 +386,184 @@ __global__ void modify_weights_ssf_kernel_1d(
...
+      weights_device[ipt] *= parent_weight / sum;
+    }
+
+    __syncwarp();
+
+  }
+}
+
 
+void cuda_reciprocal(size_t length, double* vec, cudaStream_t stream) {
```
wavefunction91, line 556: I would add this into one of the cuda utility files: not really specific to the weights (though it is the only place that its used)

src/integrator/cuda/xc_cuda_data.cxx
```diff
@@ -49,12 +51,13 @@ XCCudaData<F>::XCCudaData( size_t _natoms,
...
   buffer_adaptor mem( device_ptr, fill_sz );
 
+  LDatoms = util::div_ceil( natoms, cuda::weight_unroll ) * cuda::weight_unroll;
 
   shells_device     = mem.aligned_alloc<Shell<F>>( nshells );
   exc_device        = mem.aligned_alloc<F>( 1 );
   nel_device        = mem.aligned_alloc<F>( 1 );
   acc_scr_device    = mem.aligned_alloc<F>( 1 );
-  rab_device        = mem.aligned_alloc<F>( natoms * natoms );
+  rab_device        = mem.aligned_alloc<F>( LDatoms * natoms, 2 * sizeof(F));
```
wavefunction91, line 60: This aligns the atoms across access boundaries, yes?

wavefunction91 (changes requested):
Overall looks great!

src/integrator/cuda/cuda_device_properties.hpp
```diff
@@ -10,6 +10,14 @@ static constexpr uint32_t max_warps_per_thread_block =
   max_threads_per_thread_block / warp_size;
 
 static constexpr uint32_t max_submat_blocks = 10;
+
+// Properties for weight algorithm
+static constexpr uint32_t weight_unroll = 4;
+static_assert(weight_unroll == 4, "Weight unroll is only tested for value of 4");
+static constexpr uint32_t weight_thread_block = 640;
```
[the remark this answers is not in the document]
dmclark17: A thread block size of 640 allows for 63% occupancy. There are other sizes which allow for this, however I found this performed the best.

I also found that 4 was the optimal unroll factor through experimentation. A lower unroll factor gives lower register usage and more efficient early exiting, but a higher unroll factor hides the memory latency better.

src/integrator/cuda/xc_cuda_data.cxx
dmclark17: Yes, this is padding the number of atoms and aligning the memory to 16 bytes, which is required for double2 vector loads.

I do assume that LDatoms is even in order for each row of rab_device to be aligned to 16 bytes. That is part of the reason for the static assert for `weight_unroll`.

src/integrator/cuda/cuda_extensions.hpp
```diff
@@ -29,6 +29,13 @@ double warpReduceSum(double val) {
   return val;
 }
 
+__inline__ __device__
+double warpReduceProd(double val) {
```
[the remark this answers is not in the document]
dmclark17: I believe that the compiler is already unrolling the loop

src/integrator/cuda/cuda_weights.cu
dmclark17: Is there a .cu utility file to put the kernel itself in? Or is it okay if that stays in the weights file? If not, I can make one

src/integrator/cuda/xc_cuda_data.cxx
wavefunction91: OK, I think that works in general. I guess that setting it explicitly to align to 16 bytes (or `sizeof(double2)`) might be a bit more clear, just in case we need to look back at this in the future and wonder why a factor of 2 magically appears

src/integrator/cuda/cuda_device_properties.hpp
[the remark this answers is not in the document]
wavefunction91: Hmm, ok. I think it should be fine to leave as is for now (noting that this will not generally be portable across systems with a different SM count / max block size, etc). We should probably add a configure file somewhere that sets these for specific hardware (since they are empirically measured)

src/integrator/cuda/cuda_weights.cu
wavefunction91: I would probably move this to `xc_cuda_util`, but we may need to do some refactoring in a bit anyways, so we can probably hold off for a bit on this

src/integrator/cuda/cuda_weights.cu
wavefunction91: Do we know if `nvcc` is smart enough to desipher / optimize around a `constexpr` constant = 32 vs just manually specifying `32` everywhere? My hope is yes, but I'm not sure I've ever confirmed this

dmclark17: I confirmed via the generated PTX that in this case `nvcc` is able to decipher the `constexpr` and optimize around it.

wavefunction91: Good to know for future reference!

wavefunction91 (approved):
This look great, thanks! I'll pull and verify everything works on Summit and merge ASAP

PR #44, opened 2021-05-20
Handle CUDA Architectures

### LINKED ISSUE #17: [CUDA] Handle CUDA Architectures
Currently, we don't pass or determine any specific flags for CUDA architectures. Simplest way to handle this is to bump CMake -> 3.18 and enforce a minimum compute capability via `CMAKE_CUDA_ARCHITECTURES` if not set. The use of FP64 `atomicAdd` should put the minimum compute capability at `60`

wavefunction91:
closes #17


wavefunction91:
@evaleev Does CMake 3.18 cause problems downstream in MQPC?

Issue #56, opened 2023-07-28
nprim_pair_max is too small


dmejiar:
The parameter `nprim_pair_max` is too small for many combinations of atoms and basis sets. For example, C/cc-pvqz, Ar/cc-pvdz, Fe/cc-pvdz will lead to the following error:

```
terminate called after throwing an instance of 'GauXC::generic_gauxc_exception'
  what():  Generic GauXC Exception (Too Many Primitive Pairs)
  File     /Users/meji656/Sources/exachem-dev/build/TAMM_External-prefix/src/TAMM_External-build/GauXC_External-prefix/src/GauXC_External/include/gauxc/shell_pair.hpp
  Function void GauXC::ShellPair<F>::generate(const_shell_ref, const_shell_ref) [with F = double; const_shell_ref = const GauXC::Shell<double>&]
  Line     70
```

wavefunction91:
We are aware of this restriction and are working on a performant fix. Unfortunately, simply setting `nprim_pair_max` to a large value will trigger poor memory usage in the GPU kernels (as it's compile time). If you're focusing on just getting something to run on CPUs, we can provide an override in the interim.

PR #57, opened 2023-08-02
Temporary fix for `nprim_pair_max` Override

wavefunction91:
Adds an (undocumented) `GAUXC_SHELLPAIR_PRIMPAIR_MAX` configure variable to modulate `nprim_pair_max`. This is a temporary solution for #56 until a more permanent/performant solution can be put in place. Setting this parameter > 64 with GPU sn-K will likely lead to performance degradation and potential runtime errors.


wavefunction91:
@dmejiar Please let me know if this fixes your problem in #56 .

dmejiar:
Thanks @wavefunction91 . I think this helps. What do you think @ajaypanyala ?

ajaypanyala:
@wavefunction91 @dmejiar I have tested it and it works. Thanks!

PR #72, opened 2023-08-22
Libxc should be cited alongside GauXC

susilehtola:
The GauXC citations don't include Libxc at the moment. Libxc should be included in the list of citations to make sure people don't cite just GauXC...

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ If GauXC was used for the evaluation of exchange-correlation related
 quantities in your publication, we request that you also cite
 [Libxc](https://libxc.gitlab.io/) which provides the underlying
 implementation of the exchange-correlation functionals used in GauXC
-(via the ExchCXX library):
+via the [ExchCXX](https://github.com/wavefunction91/ExchCXX) library:
 
 ```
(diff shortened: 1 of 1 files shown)


wavefunction91 (changes requested):
I'm generally supportive of this, but I would put this as a separate block below the GauXC citations. While Libxc provides an important functionality for a subset of GauXC, indicating that it should be cited for *any* GauXC usage (e.g. even for sn-K) does not seem appropriate.

Something to the effect of 
"If GauXC was used for the evaluation of exchange-correlation related quantities in your publication or software, we request that you additionally cite Libxc (link to Gitlab) which provides the underlying implementation of the exchange-correlation functionals used in GauXC via ExchCXX:"
<Insert your BibTeX here>

In addition, we should clarify some of this in ExchCXX as well, please open an *Issue* in ExchCXX to track, I will work on verbiage and ping you for review.

wavefunction91 (changes requested):
Other than that, g2g. Thanks for the changes

susilehtola:
Go ahead and merge

wavefunction91 (approved):

PR #78, opened 2023-10-24
Fix UKS Input

### LINKED ISSUE #76: UKS inputs require non-canonical scaling factors
Currently, the UKS API has the following calling signature
```
[EXC, VXC_S, VXC_Z] = INTEGRATOR(D_S, D_Z);
```
where $D_s = \frac{1}{2}(D_\alpha + D_\beta)$ and $D_z = \frac{1}{2}(D_\alpha - D_\beta)$, etc. Although these are the formally correct factors in the Pauli expansion

$$
D^{UKS} = D_s \otimes I_2 + D_z \otimes \sigma_z
$$

it is more canonical to externally handle the additional factors of $\frac{1}{2}$ such that $D_s$ and $D_z$ carry the physical interpretations of the total- and spin-densities, respectively. We should also use this convention to avoid unnecessary confusion (HT @ajaypanyala)

wavefunction91:
Prior to this PR, the UKS drivers took $\frac{1}{2}(D_s \pm D_z)$ as their input. This PR removes the factor of $\frac{1}{2}$ such that the inputs retain some semblance of physical interpretation. N.B. the outputs of the UKS drivers remain the same, as it turns out that the factors cancel in any (equivalent) definition of the input densities / output potentials. See discussion in #76 + [these notes](http://wavefunction91.github.io/_assets/spinor_dft.pdf) for appropriate derivations.

Closes #76


wavefunction91:
Ping @dmejiar and @elambros, please confirm that this produces the expected input/outpur behaviour in `ExaChem` and `ChronusQ`.

ajaypanyala:
@wavefunction91 Confirming that this PR works fine after removing the 1/2 in our code.

PR #80, opened 2023-10-25
Unify RKS and UKS Implementations

### LINKED ISSUE #60: Unify Internal Drivers for RKS and UKS
The UKS implementation introduced in #59 replicates most of the code used for the RKS integrator. While the user APIs need to be separated (as implemented), the internal implementation can be unified by checking for e.g.-`nullptr` / zero-dims in the passed data pointers.

wavefunction91:
Builds upon #78. Unify RKS and UKS implementations + various API fixes and documentation additions. Closes #60


wavefunction91:
@elambros This is the PR GKS should be built upon

ajaypanyala:
@wavefunction91 Confirming that this works.

PR #86, opened 2023-11-16
GKS

elambros:
Adds in a basic implementation of generalized Kohn Sham for X2C-DFT calculations on HOST following: https://doi.org/10.1140/epjb/e2018-90170-1


src/xc_integrator/local_work_driver/host/reference_local_host_work_driver.cxx
```diff
@@ -141,7 +141,55 @@ namespace GauXC {
...
+
+      const double rhos = blas::dot( nbe, basis_eval + ioffs, 1, Xs_i, 1 );
+      const double rhoz = blas::dot( nbe, basis_eval + ioffz, 1, Xz_i, 1 );
+      const double rhox = blas::dot( nbe, basis_eval + ioffx, 1, Xx_i, 1 );
+      const double rhoy = blas::dot( nbe, basis_eval + ioffy, 1, Xy_i, 1 );
+ 
+      double mtemp = rhoz * rhoz + rhox * rhox + rhoy * rhoy;
+      double mnorm = 0;
+
+      if (mtemp > 1.0e-24) {
```
wavefunction91, line 173: I dislike "magic numbers". We should somehow make this configurable (preferably runtime, but I would accept a `constexpr` variable just so we can put a name on it

src/xc_integrator/replicated/host/reference_replicated_xc_host_integrator_exc_vxc.hpp
```diff
@@ -200,11 +288,17 @@ void ReferenceReplicatedXCHostIntegrator<ValueType>::
     const int32_t* shell_list = task.bfn_screening.shell_list.data();
 
     // Allocate enough memory for batch
+     
 
-    const size_t spin_dim_scal = is_rks ? 1 : 2; 
-    // Things that every calc needs
+    size_t spin_dim_scal_temp = is_rks ? 1 : 2;
+    spin_dim_scal_temp             *= is_gks ? 2 : 1;
+    const size_t spin_dim_scal = spin_dim_scal_temp;
```
wavefunction91, line 295:
```suggestion
    const size_t spin_dim_scal = is_rks ? 1 : is_uks ? 2 : 4; // last case is_gks
```

src/xc_integrator/replicated/host/reference_replicated_xc_host_integrator_exc_vxc.hpp
```diff
@@ -292,13 +403,25 @@ void ReferenceReplicatedXCHostIntegrator<ValueType>::
...
       if(is_rks) {
         lwd->eval_uvvar_lda_rks( npts, nbe, basis_eval, zmat, nbe, den_eval );
       } else if(is_uks) {
         lwd->eval_uvvar_lda_uks( npts, nbe, basis_eval, zmat, nbe, zmat_z, nbe,
           den_eval );
+      } else if(is_gks) {
+        lwd->eval_uvvar_lda_gks( npts, nbe, basis_eval, zmat, nbe, zmat_z, nbe,
+          zmat_x, nbe, zmat_y, nbe, den_eval, K );
+      } else {
+        GAUXC_GENERIC_EXCEPTION("MUST BE EITHER RKS, UKS, or GKS!");
```
wavefunction91, line 424: I would move this exception out to where you initially populate `is_{r,u,g}ks`

wavefunction91 (changes requested):
Need to add unit tests, I'll leave you to decide what's something reasonable. The only thing I might suggest is that you add something with a non-trivial magnetization orientation just to make sure that we're handling the `x,y,z` components correctly.

src/xc_integrator/local_work_driver/host/reference_local_host_work_driver.cxx
elambros: Yep added in runtime configurability for this

include/gauxc/xc_integrator_settings.hpp
```diff
@@ -16,4 +16,9 @@ struct IntegratorSettingsSNLinK : public IntegratorSettingsEXX {
   double k_tol      = 1e-10;
 };
 
+struct IntegratorSettingsEXCVXC { virtual ~IntegratorSettingsEXCVXC() noexcept = default; };
```
wavefunction91, line 19:
```suggestion
struct IntegratorSettingsXC { virtual ~IntegratorSettingsXC() noexcept = default; };
```
I can think of a few things we can put here that will propagate to a number of different places. Best to leave it a generic name.

src/xc_integrator/local_work_driver/host/reference_local_host_work_driver.cxx
wavefunction91: I see that, thanks! I've been meaning to add a set of RT parameters for XC, just hadn't gotten around to it! See updated review comment.

include/gauxc/xc_integrator_settings.hpp
elambros: Agreed, made that change.

include/gauxc/xc_integrator/impl.hpp
```diff
@@ -45,6 +45,13 @@ typename XCIntegrator<MatrixType>::exc_vxc_type_uks
   return pimpl_->eval_exc_vxc(Pscalar, Pz);
 };
 
+template <typename MatrixType>
+typename XCIntegrator<MatrixType>::exc_vxc_type_gks
+  XCIntegrator<MatrixType>::eval_exc_vxc( const MatrixType& Pscalar, const MatrixType& Pz, const MatrixType& Px, const MatrixType& Py ) {
```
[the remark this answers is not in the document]
elambros: Adressed (1) 

For (2) in CQ these don't get passed around as separate args, we just pass in the matrix object that has X,Y, and Z components. If using those, we check first if "has_Z" and then "has_XY",  so from that I got ZXY.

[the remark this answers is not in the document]
wavefunction91: OK, so then it doesn't matter on your end. If it's all the same, I'd prefer to have the ordering swapped.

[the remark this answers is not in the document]
elambros: Sounds good. Ordering has been switched to ZYX

wavefunction91 (changes requested):
Code looks good, but CI is failing
https://github.com/wavefunction91/GauXC/actions/runs/6909344016/job/18801013197

wavefunction91 (approved):

PR #95, opened 2024-02-17
Handle external gau2grid

### LINKED ISSUE #93: flexible dependency detection
Background: Most Psi4 builds (by devs, users, and packagers) detect pre-built installations of its dependency packages, rather than building them with FetchContent (hereafter FC), including gau2grid (g2g).

* From email discussions, it sounds like gauxc includes a pre-generated g2g source and expects to build that through FC. It'd be handy to have the option to detect a prebuilt and to not have two g2g libraries running around.
* Psi4 has no problems requiring CMake ~3.24 to use `FetchContent(... FIND_PACKAGE)` if that'll help.
* I can change any CMake for the g2g repo if that'll help.

wavefunction91:
By default, GauXC expects to handle gau2grid via a pre-generated source (unless forced with `GAUXC_FORCE_EXTERNAL_GAU2GRID`). It is desirable to handle the case where gau2grid is built in an external tree (#93).

Closes #93

### DIFF
--- a/cmake/gauxc-gau2grid.cmake
+++ b/cmake/gauxc-gau2grid.cmake
@@ -1,36 +1,41 @@
 if( GAUXC_ENABLE_GAU2GRID )
-
-if( GAUXC_FORCE_EXTERNAL_GAU2GRID )
-
-  find_package( gau2grid CONFIG QUIET ) 
-  if( NOT gau2grid_FOUND )
-  
-    include( gauxc-dep-versions )
-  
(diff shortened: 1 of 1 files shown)


davpoolechem:
@loriab and I worked today on testing this PR within the context of the Psi4/GauXC interface. It seems to work as desired, both in terms of finding the external gau2grid instance, and in subsequent accuracy tests. Just wanted to say that!

PR #103, opened 2024-03-26
Fix CMake Variable Naming

### LINKED ISSUE #102: CMake Variable Naming Concerns
In light of discussion present in https://github.com/wavefunction91/GauXC/pull/101, it seems that there may be some concerns with the naming scheme used for certain GauXC variables within the CMake build system, which is relevant since there is intent to export some of these variables as properties within the context of a CMake build. The current scheme of `GAUXC_ENABLE_X`, as pointed out by @loriab, implies that the variable is editable by the end user.

wavefunction91:
Previously, the `GAUXC_ENABLE_XYZ` variables did all the heavy lifting for feature detection and control. Per @loriab's comments and subsequent issue (#102),these variables sound suspiciously editable, so probably best to not rely on that naming scheme.

This PR separates out `GAUXC_ENABLE_XYZ` to trigger detection of  feature `XYZ` and sets `GAUXC_HAS_XYZ` if the feature has been detected and enabled throughout the stack.

**Major Change: Do not use `GAUXC_ENABLE_XYZ` downstream to check feature behaviour, only use these variables to enable features via FetchContent and use `GAUXC_HAS_XYZ` to check if the feature enable happened. The latter should also be used for source-level feature detection (e.g. `GAUXC_HAS_DEVICE` can be used to check if GauXC has device bindings)**

Closes #102 

TODO: Check downstream integrations:
- [x] NWChemEx
- [x] MPQC
- [x] ExaChem
- [x] ChronusQ
- [x] Psi4


wavefunction91:
Ping @evaleev @dmejiar @ajaypanyala @elambros @loriab @davpoolechem 

This will undoubtedly break some things in some build systems. Please check that you can make this work in your respective code bases. I won't merge until we can get consensus that we're not breaking things downstream.

ajaypanyala:
`ExaChem` is good to go!

wavefunction91:
MPQC is fixed as of https://github.com/ValeevGroup/mpqc4/pull/464 @evaleev.

davpoolechem:
Can confirm that this works for Psi4!

wavefunction91:
NWChemEx integration is fixed

wavefunction91:
@elambros can you confirm that this PR works in CQ? It will require changing all instances of `GAUXC_ENABLE_DEVICE` to `GAUXC_HAS_DEVICE` for the device feature check in the code (fetch/configure should stay the same)

elambros:
Yep works on the CQ end too!

PR #120, opened 2024-04-08
Semi-automatic collocation code generation on device

mikovtun:
Adds an additional pyexpander script to complete collocation code generation up to arbitrary angular momentum.

### DIFF
--- a/src/xc_integrator/local_work_driver/device/cuda/kernels/collocation/templates/collocation_device_template.cu
+++ b/src/xc_integrator/local_work_driver/device/cuda/kernels/collocation/templates/collocation_device_template.cu
@@ -1,0 +1,484 @@
+/**
+ * GauXC Copyright (c) 2020-2024, The Regents of the University of California,
+ * through Lawrence Berkeley National Laboratory (subject to receipt of
+ * any required approvals from the U.S. Dept. of Energy). All rights reserved.
+ *
+ * See LICENSE.txt for details
+ */
+#include <gauxc/util/div_ceil.hpp>
+#include "device_specific/cuda_util.hpp"
+#include "exceptions/cuda_exception.hpp"
+#include <gauxc/xc_task.hpp>
+
+#include "device/common/collocation_device.hpp"
+#include "device/cuda/kernels/collocation_masked_kernels.hpp"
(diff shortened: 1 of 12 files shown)


src/xc_integrator/local_work_driver/device/cuda/kernels/collocation/templates/collocation_device_template.cu
```diff
@@ -0,0 +1,484 @@
...
+
+
+}
+
+
+uint32_t max_threads_shell_to_task_collocation_hessian( int32_t l, bool pure ) {
+  if( pure ) {
+    switch(l) {
+      case 0: return util::cuda_kernel_max_threads_per_block( collocation_device_shell_to_task_kernel_cartesian_hessian_0 );\
+      $for( L in range(1, L_max) )
```
wavefunction91, line 408:
```suggestion
      $for( L in range(1, L_max + 1) )
```

src/xc_integrator/local_work_driver/device/cuda/kernels/collocation/templates/collocation_device_template.cu
```diff
@@ -0,0 +1,484 @@
...
+      case $(L): return util::cuda_kernel_max_threads_per_block( collocation_device_shell_to_task_kernel_spherical_hessian_$(L) );\
+      $endfor
+      default: GAUXC_GENERIC_EXCEPTION("CUDA L_MAX = $(L_max)");
+    }
+  } else {
+    switch(l) {\
+      $for( L in range(L_max) )
```
wavefunction91, line 415:
```suggestion
      $for( L in range(L_max + 1) )
```

wavefunction91 (changes requested):
Add the files generated by these new scripts (for `LMAX = 4`) to this PR - there's some spacing oddities that are artifacts of `pyexpander` - they're fine, but it would be nice to have an accounting of when strange things like this get introdced. Other than that. LGTM - nice QoL improvement.

You're also more than welcome to bump `LMAX` in a separate PR - the only thing is that if you want to do that, you'd need to add some tests to make sure we get coverage. Not a huge deal for right now

wavefunction91 (approved):

wavefunction91 (approved):

PR #132, opened 2024-06-13
[CMake][CUDA] static libs

ajaypanyala:


attic/src/integrator/cuda/gauxc-cuda_integrator.cmake
```diff
@@ -64,7 +64,12 @@ else()
 
 endif()
 
-target_link_libraries( gauxc PUBLIC CUDA::cublas )
+if(BUILD_SHARED_LIBS)
```
wavefunction91, line 67: Per offline comments - the way GauXC is build (shared/static) is orthogonal to how CUDA is linked. I agree exposing this flexibility is important (because you can't/shouldn't link static + shared CUDA in the same project), but it should default based on `BUILD_SHARED_LIBS`. I would rename this `GAUXC_LINK_CUDA_STATIC`, defaulted to `OFF` and in the top-level project setting
```
set(GAUXC_LINK_CUDA_STATIC ${BUILD_SHARED_LIBS} CACHE BOOL "Link GauXC with static CUDA libs" FORCE)
```

wavefunction91 (changes requested):

wavefunction91 (approved):

wavefunction91 (approved):

Issue #134, opened 2024-07-08
Numerical inaccuracies on consumer grade GPUs


mikovtun:
Running the test suite on device (#91) fails for the H3 test system (GKS GGA) when running on an NVIDIA RTX 3060 because the calculated `VXCx` and `VXCy` norms differ from the CPU calculations by about `1e-7`. This failure does not occur on the professional line of cards (tested on A100, L40, and A40). The likely cause is a different implementation of the `rsqrt` call between the RTX and professional cards. 

This is the first time different GPUs have caused a non-negligible difference in calculations, which this issue serves to document. While consumer cards can be used (with caution), they are not the intended hardware to be used with GauXC.

PR #91, opened 2023-11-28
Device UKS/GKS Implementation

mikovtun:
This PR adds device (CUDA) support for XC calculations with Unrestricted and Generalized Kohn-Sham references. LDA and GGA functionals are supported.

PR #135, opened 2024-07-18
HIP compilation

ryanstocks00:
Similar to https://github.com/wavefunction91/ExchCXX/pull/39, I had to make these minor changes to get GauXC to compile with HIP

### DIFF
--- a/src/xc_integrator/local_work_driver/device/scheme1_base.cxx
+++ b/src/xc_integrator/local_work_driver/device/scheme1_base.cxx
@@ -441,9 +441,9 @@ void AoSScheme1Base::eval_collocation_hessian( XCDeviceData* _data ) {
   eval_collocation_shell_to_task_hessian( max_l, 
     data->l_batched_shell_to_task.data(), aos_stack.device_tasks,
     data->device_backend_->queue() );
-#endif
   
   data->device_backend_->check_error("collocation hess" __FILE__ ": " + std::to_string(__LINE__));
+#endif
 }
 
 void AoSScheme1Base::eval_collocation_laplacian( XCDeviceData* _data ) {
@@ -461,9 +461,9 @@ void AoSScheme1Base::eval_collocation_laplacian( XCDeviceData* _data ) {
   eval_collocation_shell_to_task_laplacian( max_l, 
     data->l_batched_shell_to_task.data(), aos_stack.device_tasks,
     data->device_backend_->queue() );
-#endif
   
   data->device_backend_->check_error("collocation lapl" __FILE__ ": " + std::to_string(__LINE__));
+#endif
 }
 
 


src/xc_integrator/local_work_driver/device/cuda/kernels/uvvars.cu
```diff
@@ -306,7 +306,7 @@ void eval_uvvars_gga( size_t ntasks, size_t npts_total, int32_t nbf_max,
   {
   dim3 threads( cuda::warp_size, cuda::max_warps_per_thread_block / 2, 1 );
   dim3 blocks( std::min(uint64_t(4), util::div_ceil( nbf_max, 4 )),
-               std::min(uint64_t(16), util::div_ceil( nbf_max, 16 )),
+               std::min(uint64_t(GGA_KERNEL_SM_BLOCK_Y), util::div_ceil( npts_max, GGA_KERNEL_SM_BLOCK_Y )),
```
ryanstocks00, line 309: I think the nbf_max usage here was potentially a bug? Have replaced with nbf_max

wavefunction91: It's not a bug, but potentially not optimal on some hardware. What have you been testing on? In principle, these parameters should be tuned, these were just the ones that were found to perform best on V100/A100.

wavefunction91 (commented):
This is a bit more complicated than https://github.com/wavefunction91/ExchCXX/pull/39, I'm happy to work with you on getting this fixed up, but the HIP implementation is GauXC is a bit delicate (or at least it was for older HIP/ROCm on MI250/300 a few years ago when this was a priority) - the kernels are prone to silent errors due to suboptimal resource usage emitted by the optimizing compilers (i.e. the kernel is too large for the launch parameters, and we never really hardened the implementation to do the "right" thing). Getting e.g. the `uvvars` kernels to work was quite a bit of effort, IIRC.

I can provide you with some of our canonical stress test cases to validate the implementation - again, I don't have access to AMD hardware at the moment, so I can't really test anything on my end.

ajaypanyala:
@wavefunction91 @ryanstocks00 After a couple of minor tweaks to the build to enable successful `hipblas` discovery, I was able to build the code on Frontier.  However, when I run a test, I get 

```terminate called after throwing an instance of 'GauXC::generic_gauxc_exception'
  what():  Generic GauXC Exception (EXX + non-CUDA NYI)
  File     /lustre/orion/chm216/scratch/panyala/TAMM/bdft/GauXC_External-prefix/src/GauXC_External/src/xc_integrator/local_work_driver/device/scheme1_base.cxx
  Function virtual void GauXC::AoSScheme1Base::eval_exx_ek_screening_bfn_stats(XCDeviceData *)
  Line     1389
```

wavefunction91:
Thanks Ajay, yea, we don't have a HIP implementation of snK (although, it might be easy to add, would just need to tune the kernel params, it's linear along the wavefront), so we just need to add the logic to not run that test.

Can you provide the diff for the build system stuff?

ajaypanyala:
No worries, I realized that just now reg. sn-K. The regular XC eval works fine. Can I commit the build system changes to this PR ?

wavefunction91:
That would be up to @ryanstocks00.

src/xc_integrator/local_work_driver/device/cuda/kernels/uvvars.cu
ryanstocks00: Is it guaranteed that npts_max is greater than nbf_max? My understanding from a perusal of the function is that the y axis is iterating over the points rather than basis functions so the nbf_max could have been problematic? I'm still working on testing it all out, so certainly haven't got as far as performance tuning yet

ryanstocks00:
> This is a bit more complicated than [wavefunction91/ExchCXX#39](https://github.com/wavefunction91/ExchCXX/pull/39), I'm happy to work with you on getting this fixed up, but the HIP implementation is GauXC is a bit delicate (or at least it was for older HIP/ROCm on MI250/300 a few years ago when this was a priority) - the kernels are prone to silent errors due to suboptimal resource usage emitted by the optimizing compilers (i.e. the kernel is too large for the launch parameters, and we never really hardened the implementation to do the "right" thing). Getting e.g. the `uvvars` kernels to work was quite a bit of effort, IIRC.
> 
> I can provide you with some of our canonical stress test cases to validate the implementation - again, I don't have access to AMD hardware at the moment, so I can't really test anything on my end.

Yes we have had similar issues on the MI250 hardware. My main priority here was getting something that would compile so as not to break our HIP CI pipeline. I haven't yet gone through thorough testing as I was going to get it integrated on nvidia systems first. That said we do have access to AMD hardware so am happy to do whatever testing you think will be useful. I'm hopeful there will be more useful contributions going forward as I wrap my head around it all.

@ajaypanyala happy for you to commit changes to this PR - do you have the required permissions to push to the branch in my repo?

src/xc_integrator/local_work_driver/device/cuda/kernels/uvvars.cu
wavefunction91: It's not guaranteed, although it usually is. I agree, what's there is likely a typo, but the kernel is hardened to take any block/grid dimension and still give the right results (i.e. whether or not loops get executed is based on the number of warps in the thread block).

I'll check to see whether this kneecaps the performance of this kernel in prod - if there's no change or it's better, I'll accept it for being "correct", if it's worse I'll come back with a hand-wavy/tin-foil-hat reason for why that's the case :).

ajaypanyala:
> @ajaypanyala happy for you to commit changes to this PR - do you have the required permissions to push to the branch in my repo?
I do not have the permissions. Could you please add me ?

wavefunction91:
@ajaypanyala How are you testing this? Just the UTs or are you running something non-trivial (e.g. Ubi)? I'd like to get a full run of Ubi/DZ on AMD HW before merging this. 

Also, we might want to wait until #91 is merged as it will require additional updates to get to work with HIP - potentially some issues to work out viz runtime errors as well for large systems. Might need to coordinate with @mikovtun to get a bit test system (they have some big things they're testing with).

ajaypanyala:
@wavefunction91 Tested with Ubi/DZ (pbe0) on MI250X.

ajaypanyala:
@wavefunction91 Is this ready to go (modulo the merge conflict) ?

ryanstocks00:
@wavefunction91 @ajaypanyala I have fixed the merge conflicts so that it successfully compiles with HIP, however it fails a lot of tests, would be great to get some more experienced eyes over it

wavefunction91:
Hi @ryanstocks00, sorry for the delay. I've recently changed jobs, so I'm still in the process of renormalizing where my time is spent.

Great that this compiles. Could you provide more info on which tests are failing? That will help us pin point where things could be going wrong. It might also be worth running the standalone_driver through rocgdb to see if we can see which kernels are dying. If I recall the last time I worked with AMD/HIP, the issue was invariably silent kernel launch failures due to hard coded launch params (optimized for A100).

ryanstocks00:
@wavefunction91 hope Microsoft is treating you well! (Is this project likely to continue much development?)

The HIP code seems to be correct for LDA functionals so I think I must have stuffed something in the GGA code during the merge (not sure the MGGA code was ever fully implemented in HIP). I attempted to run rocgdb on standalone_driver as suggested using an MI250 however didn't get any errors - is this the right process? [hip_mi250_standalone_b3lyp_rocgdb.txt](https://github.com/user-attachments/files/17188229/hip_mi250_standalone_b3lyp_rocgdb.txt) The output is non-deterministic for both EXC and VXC.

I have attached the output from running `./gauxc_test` on both NVIDIA and AMD - interestingly there is some small numeric noise on the 4080 that causes some of the tests to fail but I assume this is just because it is not a datacenter GPU)

[cuda_4080s_standalone_b3lyp.txt](https://github.com/user-attachments/files/17188052/cuda_4080s_standalone_b3lyp.txt)
[cuda_4080s_test.txt](https://github.com/user-attachments/files/17188053/cuda_4080s_test.txt)
[hip_mi250_standalone.txt](https://github.com/user-attachments/files/17188054/hip_mi250_standalone.txt)
[hip_mi250_standalone_b3lyp.txt](https://github.com/user-attachments/files/17188055/hip_mi250_standalone_b3lyp.txt)
[hip_mi250_test.txt](https://github.com/user-attachments/files/17188056/hip_mi250_test.txt)

wavefunction91:
@ryanstocks00 Thanks, this is helpful to diagnosing the problem (I think). Lots to unpack here, I'll try to cover everything.

> Is this project likely to continue much development?

Yes, updated may be rolled out slower, but dev and support will continue.

> The HIP code seems to be correct for LDA functionals so I think I must have stuffed something in the GGA code during the merge

Nothing obvious from what I can see from the updates, but I agree with this assessment based on the results you've shared

> I attempted to run rocgdb on standalone_driver as suggested using an MI250 however didn't get any errors

This *should* indicate that kernels are not failing to launch (although I'm not precluding it, `rocgdb` is not the most stable software on earth)

>  is this the right process?

Looks like it

>The output is non-deterministic for both EXC and VXC.

Can you expand on this? Both for LDA and GGA? 

> 4080

Yes, this is known https://github.com/wavefunction91/GauXC/issues/134. I'm not completely sold on the use of consumer GPUs for high-performance DFT simulations (happy to be proven wrong!), so I don't think I'll be prioritizing that for some time (that, and I don't have a consumer grade GPU to test on!)




Based on the results you've shared, I think the next think to check is whether or not the LDA gradients work on AMD. If they do, we can at least preclude the collocation gradients being the problem (as well as a more extensive testing of the  batched BLAS - they'll run through similar paths). If those work, then it's in one of two kernels - the `uvvars` or the `zmat`. Since EXC looks busted for GGAs, I'd suspect (at least) the former is the cuprit. When I was designing that kernel, it was definitely tuned for NVIDIA SM structure (e.g. warp len of 32 + square process grids within the block). I'm not sure the behaviour of e.g. the warp level reductions on AMD and whether or not there's a non-obvious race condition. This should also be a problem for LDA (i.e. if the logic is fixed there, it should be fixed everywhere), but It might be worth another look.

Again, I don't have access to AMD hardware at the moment, so there's not a ton I can do on the debugging side. Happy to brainstorm debugging ideas though, let me know.






P.S. I'm sure you're aware of this, but just to be explicit - running the standalone driver with b3lyp for benzene will fail the checks as only the SVWN5 data is saved there. I'm assuming you ran those to get the reference numbers for B3LYP?

ryanstocks00:
@wavefunction91 thank you very much for the detailed response. I don't think we can check the grads on AMD as I get a "Generic GauXC Exception (LDA Grad NYI for HIP Backends)".

**The output is non-deterministic for both EXC and VXC. _Can you expand on this? Both for LDA and GGA?_**
Using the standalone driver, EXC and VXC are both correct for LDA, both non-deterministic for GGA (and order of magnitude different to the correct value calculated with the host CPU implementation)

I have finally managed to get it to compile and run with HIP on a NVIDIA machine and get much the same results as on AMD (correct LDA, non-deterministic GGA) which I think rules out AMD launch configuration issues and probably warp length issues so I'm pretty sure it is a code issue. This could potentially be a way for you to test as well since the issue is replicatable on NVIDIA hardware. If there's a good way to narrow down which kernel is the issue that would be very helpful (e.g. does the fact that the "Li / SVWN5 / sto-3g" tests in gauxc_test fail despite being LDA mean anything?).

ryanstocks00:
I think I have now got gauxc running correctly with HIP on a NVIDIA machine - the problem was the vvar grad kernel which required the thread block to be square. I think this will still be problematic on AMD machines as there isn't sufficient shared memory with the increased warp length so will need some more modification though I'm going to have to put this on the backburner for a bit due to other priorities

wavefunction91:
> `vvar_grad`

For the GGA energy or for the gradient? You should only call the `vvar` kernel for non-gradients.

Admittedly, I hadn't fully appreciated how complicated that code became with the latest refactor to support GKS, etc. I'll try to set aside some time to look at this over the coming week. FWIW - this set of kernels used to work (obviously), it might just require looking over the [AMD commits](https://github.com/wavefunction91/GauXC/commits?author=dmcdougall) to see what worked in the past and try to see how to translate it into the new format.

Thanks for continuing to take a look at this, the effort has been very helpful and much appreciated.

mikovtun:
Hey, if I can weigh in, when I was authoring the nvidia kernels I noticed a _very_ similar output when I was testing with a 3060ti. If I remember right, I was getting that very same `0.0000001368` diff on the H3 VXCs, and the culprit was nvidia's implementation of `sqrt` differing between product lines. As documented in #134, the issue goes away on an A100. Could be a coincidence you got that same value, or it could point to AMDs impl using the same algorithm for `sqrt` as the consumer grade nvidia cards. Nice PR!

ryanstocks00:
@mikovtun good pickup! I think you might be on to something - If I use CUDA on a 4080 I get exactly the same 0.0000001368 errors, however I do not get the 0.0000000131 or 0.0000000009 errors. Interestingly if I compile the HIP version for the 4080 I _do_ get the 0.0000000131 and 0.0000000009 errors so it is clearly not only a hardware discrepancy? Bizarre but at least all the errors are now replicable on NVIDIA hardware

PR #137, opened 2024-09-10
guard against gnu exetension

JorgeG94:
#warning is a GNU extension and this fails to compile with NVHPC, I added a little guard :)

### DIFF
--- a/src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
+++ b/src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
@@ -252,9 +252,11 @@ namespace XCPU {
 
 // Scalar SIMD Emulation
 #else
-
+#ifdef __GNUC__
   #warning "Warning: ISA Not Specified: Using Scalar Code"
-
+#else
+  #pragma message "Warning: ISA Not Specified: Using Scalar Code"
+#endif
   #define SIMD_TYPE double
   
   #define SIMD_LENGTH 1


wavefunction91 (approved):
lgtm thanks - however, I've not ever had a problem with NVHPC, strange that this is a problem now. Actually, I've also never had a problem LLVM compilers in general (it's in the [CI pipeline](https://github.com/wavefunction91/GauXC/blob/master/.github/workflows/build_and_test_compiler_zoo.yml#L17)...).

Can you share versions / environment, etc? 

Also - if you're seeing this warning, you're likely compiling without vectorization. This should be avoided, as the host part of the GPU code path is sensitive to this as well. In general, if you're compiling in `Release`, you should also append

```-DCMAKE_CXX_FLAGS="-march=native"```

we've gone back and forth about adding this by default (ala Psi4's build harness), but discovered a few edge cases where that might not be optimal

JorgeG94:
> lgtm thanks - however, I've not ever had a problem with NVHPC, strange that this is a problem now. Actually, I've also never had a problem LLVM compilers in general (it's in the [CI pipeline](https://github.com/wavefunction91/GauXC/blob/master/.github/workflows/build_and_test_compiler_zoo.yml#L17)...).
> 
> Can you share versions / environment, etc?
> 
> Also - if you're seeing this warning, you're likely compiling without vectorization. This should be avoided, as the host part of the GPU code path is sensitive to this as well. In general, if you're compiling in `Release`, you should also append
> 
> `-DCMAKE_CXX_FLAGS="-march=native"`
> 
> we've gone back and forth about adding this by default (ala Psi4's build harness), but discovered a few edge cases where that might not be optimal

Hi! this was with nvhpc 24.7 on a grace hopper system. I've had weird experiences with nvhpc complaining about code when other compilers do not. I tried adding march=native to my compilation and it still found the #warning (when I reverted my small change). I just build with :  `cmake -DCMAKE_CXX_FLAGS="-march=native" ../`

PR #145, opened 2025-05-14
Fixes Molecule Ctors

ryanmrichard:
The original templated ctor has a signature compatible with copy/move construction and a  higher precedence than the copy/move ctors. In turn, attempting to copy/move a `Molecule` object causes the compiler to pick the templated ctor, which leads to a compiler error. 

This PR:
- Introduces SFINAE to fix the titular problem.
- Adds unit tests so this doesn't break again.
- Removes the use of some deprecated C++ standard library functions.

### DIFF
--- a/include/gauxc/molecule.hpp
+++ b/include/gauxc/molecule.hpp
@@ -16,10 +16,16 @@
 namespace GauXC {
 
 class Molecule : public std::vector<Atom> {
+private:
+  /// Testes if the base class can be constructed from @p Args
+  template <typename... Args>
+  using can_construct_base_t = 
+    std::is_constructible_v<std::vector<Atom>, Args...>;
 
 public:
 
-  template <typename... Args>
+  template <typename... Args, 
+            typename = std::enable_if_t<can_construct_base_t<Args...>>>
   Molecule( Args&&... args ) :
     std::vector<Atom>( std::forward<Args>(args)... ) { }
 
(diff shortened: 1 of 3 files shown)


include/gauxc/molecule.hpp
```diff
@@ -7,42 +7,57 @@
...
-    std::vector<Atom>( std::forward<Args>(args)... ) { }
+  using disable_if_molecule_t =
+      std::enable_if_t<all_not_molecule<Args...>::value>;
+
+public:
+  template <typename... Args,
+            typename = disable_if_molecule_t<std::decay_t<Args>...>>
+  Molecule(Args &&...args) : std::vector<Atom>(std::forward<Args>(args)...) {}
 
-  Molecule( const Molecule& )          = default;
```
wavefunction91, line 26: Please remove linting artifacts. I won't point them out everywhere, but there's quite a few.

wavefunction91 (changes requested):

include/gauxc/molecule.hpp
ryanmrichard: That's what I get for not looking at the diffs before submitting. Will do! I probably won't get to it until tomorrow morning.

ryanmrichard: Fixed.

wavefunction91 (approved):
Lgtm, thanks!

PR #147, opened 2025-05-20
Fixes the BasisSet class's ctor (and warnings)

ryanmrichard:
The fixes to the `BasisSet` class are analogous to the ones needed for the `Molecule` class (i.e., ensure the copy/move ctors do not try to use the templated ctor). I also went through and removed the warnings that my compiler was raising. GauXC should now be "warning free" (completely warning free will be impossible with Obara-Saika preprocessor message)

### DIFF
--- a/src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
+++ b/src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
@@ -252,11 +252,6 @@ namespace XCPU {
 
 // Scalar SIMD Emulation
 #else
-#if (defined(__GNUC__) && __cplusplus >= 202302L)
-  #warning "Warning: ISA Not Specified: Using Scalar Code"
-#else
-  #pragma message "Warning: ISA Not Specified: Using Scalar Code"
-#endif
   #define SIMD_TYPE double
   
   #define SIMD_LENGTH 1
--- a/src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
+++ b/src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
@@ -252,7 +252,7 @@ namespace XCPU {
 
 // Scalar SIMD Emulation
 #else
-#ifdef __GNUC__
+#if (defined(__GNUC__) && __cplusplus >= 202302L)
   #warning "Warning: ISA Not Specified: Using Scalar Code"
 #else
   #pragma message "Warning: ISA Not Specified: Using Scalar Code"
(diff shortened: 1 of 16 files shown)


src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
```diff
@@ -252,7 +252,7 @@ namespace XCPU {
 
 // Scalar SIMD Emulation
 #else
-#ifdef __GNUC__
+#if (defined(__GNUC__) && __cplusplus >= 202302L)
```
wavefunction91, line 255: Not sure I understand the `__cplusplus` guard here, is there a compiler this was erroring out for without it. Honestly, we can remove this warning, it was more for internal use anyways. However, you *can* make it go away by ensuring that the host compilation uses arch specific flags (e.g. `-march=native`). I don't include them by default because there's not a good (portable) way to do it, but if you're seeing this warning for a `Release` build, you should rethink how you're compiling.

src/xc_integrator/local_work_driver/host/reference_local_host_work_driver.cxx
```diff
@@ -299,7 +299,7 @@ void ReferenceLocalHostWorkDriver::eval_uvvar_mgga_rks( size_t npts, size_t nbe,
   const double* basis_eval, const double* dbasis_x_eval,
   const double *dbasis_y_eval, const double* dbasis_z_eval, const double* lbasis_eval,
   const double* X, size_t ldx, const double* mmat_x, const double* mmat_y, 
-  const double* mmat_z, size_t ldm,
+  const double* mmat_z, size_t /*ldm*/,
```
wavefunction91, line 302: This is actually a bug that should be addressed separately - we're not using the `ldm` but we should be to act as intended. Thanks for pointing it out. It would be a big help if you could propagate this, but if not, I can put it on my list.

src/xc_integrator/local_work_driver/host/reference_local_host_work_driver.cxx
```diff
@@ -1166,7 +1166,7 @@ void ReferenceLocalHostWorkDriver::eval_zmat_gga_vxc_gks( size_t npts, size_t nb
       cou_offsets_map[shell_list[i]] = cou_cart_sizes[i];
     }
 
-    size_t ndo = 0;
+    //size_t ndo = 0;
```
wavefunction91, line 1169: We can just delete this variable

wavefunction91 (changes requested):

src/xc_integrator/local_work_driver/host/obara_saika/src/config_obara_saika.hpp
ryanmrichard: I'm compiling with Apple Clang 17 and it complains that `#warning` is a C++23 feature. The `__cplusplus` guard checks that the C++ standard is C++23 or greater. That said, I'll just remove it.

src/xc_integrator/local_work_driver/host/reference_local_host_work_driver.cxx
ryanmrichard: While I can guess what `ldm` is supposed to do from looking at other functions, I don't know for sure. So I don't feel comfortable making the change on my own.

wavefunction91: No problem, that's fine, I'll take it as an action item - I know this is annoying, but could you revert the changes to this file just keep it visible?

ryanmrichard: I kept the removal of `ndo`, but reverted the changes that commented out the unused function parameters.

wavefunction91 (approved):
lgtm

PR #160, opened 2025-11-10
Allow building of shared libraries

### LINKED ISSUE #159: Cannot build GauXC shared library
Due to explicit STATIC attribute GauXC does not allow to build shared libraries if `-DBUILD_SHARED_LIBS=ON` is set:

https://github.com/wavefunction91/GauXC/blob/0319e71c5a5557e2380afd2b8a9dba7853130f41/src/CMakeLists.txt#L35-L44

It would be great if GauXC could support building shared libraries.

awvwgk:
Remove STATIC attribute from gauxc library to allow building shared libraries with `-DBUILD_SHARED_LIBS=ON`

Closes #159

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -32,7 +32,7 @@ include( gauxc-integratorxx )
 include( gauxc-exchcxx      )
 
 
-add_library( gauxc STATIC 
+add_library( gauxc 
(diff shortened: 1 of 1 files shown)


wavefunction91 (approved):
lgtm

PR #165, opened 2025-11-29
WIP: try fixing issue where GauXC insists on building its own copy of LibXC even when the Psi4 CMake has already ensured that LibXC is installed

TiborGY:
TBA


wavefunction91:
The issue with building Libxc is entirely contained in [ExchCXX](https://github.com/wavefunction91/ExchCXX) and should not require changes in GauXC. GauXC doesn't care if Libxc exists (in fact, it can be disabled entirely upstream with no downstream consequences).

TiborGY:
Yep, in fact it was more of a Psi4 issue (pinned to an old version of GauXC)

PR #174, opened 2026-01-28
Add Fortran API

### LINKED ISSUE #66: Add Fortran/C Bindings
This is a requirement for integration with legacy codes.

awvwgk:
Closes #66


src/runtime_environment.F90
```diff
@@ -0,0 +1,131 @@
...
+      type(gauxc_runtime_environment_type), intent(inout) :: rt
+    end subroutine gauxc_runtime_environment_delete
+  end interface gauxc_delete
+
+  interface
+    function gauxc_runtime_environment_comm_rank(status, rt) result(comm_size) bind(c)
+      import :: gauxc_runtime_environment_type, gauxc_status_type, c_int
+      implicit none
+      type(gauxc_status_type), intent(out) :: status
+      type(gauxc_runtime_environment_type), intent(in) :: rt
```
kjh-giesbertz: On my machine I need a copy of `rt`, so
```suggestion
      type(gauxc_runtime_environment_type), value :: rt
```

src/runtime_environment.F90
```diff
@@ -0,0 +1,131 @@
...
+      integer(c_int) :: comm_size
+    end function gauxc_runtime_environment_comm_rank
+  end interface
+
+  interface
+    function gauxc_runtime_environment_comm_size(status, rt) result(comm_size) bind(c)
+      import :: gauxc_runtime_environment_type, gauxc_status_type, c_int
+      implicit none
+      type(gauxc_status_type), intent(out) :: status
+      type(gauxc_runtime_environment_type), intent(in) :: rt
```
kjh-giesbertz: Also here I need a copy, so
```suggestion
      type(gauxc_runtime_environment_type), value :: rt
```

src/molecule.f90
```diff
@@ -0,0 +1,108 @@
...
+      !> @param status Status of the operation
+      type(gauxc_status_type), intent(out) :: status
+      !> @param molecule Pointer to the molecule object to delete
+      type(gauxc_molecule_type), intent(inout) :: molecule
+    end subroutine gauxc_molecule_delete
+  end interface gauxc_delete
+
+  interface
+    !> @brief Get the number of atoms in the molecule
+    function gauxc_molecule_get_natoms(status, molecule) result(natoms) bind(c)
```
kjh-giesbertz: This function name does not exist in the c interface, only without the "get"

```suggestion
    function gauxc_molecule_natoms(status, molecule) result(natoms) bind(c)
```

src/molecule.f90
```diff
@@ -0,0 +1,108 @@
...
+      import :: c_size_t, gauxc_status_type, gauxc_molecule_type
+      implicit none
+      !> @param status Status of the operation
+      type(gauxc_status_type), intent(out) :: status
+      !> @param molecule Pointer to the molecule object
+      type(gauxc_molecule_type), value :: molecule
+      !> @return Number of atoms in the molecule
+      integer(c_size_t) :: natoms
+    end function gauxc_molecule_get_natoms
```
kjh-giesbertz: Idem here
```suggestion
    end function gauxc_molecule_natoms
```

src/molecule.f90
```diff
@@ -0,0 +1,108 @@
...
+  use gauxc_status, only : gauxc_status_type
+  use gauxc_types, only : gauxc_header_type, gauxc_type_molecule
+  implicit none
+  private
+
+  public :: &
+    & gauxc_molecule_new, &
+    & gauxc_molecule_new_from_atoms, &
+    & gauxc_molecule_delete, &
+    & gauxc_molecule_get_natoms, &
```
kjh-giesbertz:
```suggestion
    & gauxc_molecule_natoms, &
```

src/functional.f90
```diff
@@ -0,0 +1,608 @@
...
+  end type gauxc_functional_type
+
+  interface
+    !> @brief Create a GauXCFunctional from a string specification
+    function gauxc_functional_from_string_c(status, functional_spec, polarized) &
+      & result(func) bind(c, name="gauxc_functional_from_string")
+      import :: gauxc_status_type, gauxc_functional_type, c_char, c_bool
+      type(gauxc_status_type), intent(out) :: status
+      character(kind=c_char), intent(in) :: functional_spec(*)
+      logical(c_bool), intent(in) :: polarized
```
kjh-giesbertz: This needs to be passing by copy
```suggestion
      logical(c_bool), value :: polarized
```

src/functional.f90
```diff
@@ -0,0 +1,608 @@
...
+      type(gauxc_functional_type) :: func
+    end function gauxc_functional_from_string_c
+
+    !> @brief Create a GauXCFunctional from an enum specification
+    function gauxc_functional_from_enum(status, functional_enum, polarized) &
+      & result(func) bind(c)
+      import :: gauxc_status_type, gauxc_functional_type, c_int, c_bool
+      type(gauxc_status_type), intent(out) :: status
+      integer(c_int), intent(in) :: functional_enum
+      logical(c_bool), intent(in) :: polarized
```
kjh-giesbertz: This needs to be passing by copy
```suggestion
      logical(c_bool), value :: polarized
```

src/runtime_environment.F90
```diff
@@ -0,0 +1,131 @@
...
+  end type gauxc_runtime_environment_type
+
+  interface
+    function gauxc_runtime_environment_new &
+#ifdef GAUXC_HAS_MPI
+      & (status, comm) &
+#else
+      & (status) &
+#endif
+      & result(rt) bind(c)
```
kjh-giesbertz: Make sure that the c variant is called which transforms the fortran MPI communicator
```suggestion
      & result(rt) bind(c, name="gauxc_runtime_environment_new_fortran")
```

src/c-api/c_basisset.cxx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+void gauxc_basisset_delete(GauXCStatus* status, GauXCBasisSet* basis) {
+  status->code = 0;
+  if (basis == nullptr) return;
+  if (basis->ptr != nullptr)
+    delete detail::get_basisset_ptr(*basis);
+  basis->ptr = nullptr;
+}
+
```
kjh-giesbertz, line 70: Add some additional routines to querry info in basisset object. I found it useful to check during debugging.
```suggestion

int32_t gauxc_basisset_nshells(GauXCStatus* status, const GauXCBasisSet basis) {
  status->code = 0;
  if (basis.ptr == nullptr) {
    status->code = 1;
    status->message = "No basis set associated to this object";
    return 0;
  }
  return detail::get_basisset_ptr(basis)->nshells();
}

int32_t gauxc_basisset_ngtos(GauXCStatus* status, const GauXCBasisSet basis) {
  status->code = 0;
  if (basis.ptr == nullptr) {
    status->code = 1;
    status->message = "No basis set associated to this object";
    return 0;
  }
  return detail::get_basisset_ptr(basis)->nbf();
}
```

src/fortran-api/basisset.f90
```diff
@@ -0,0 +1,76 @@
...
+      !> @param status Status of the operation
+      type(gauxc_status_type), intent(out) :: status
+      !> @param shells Pointer to an array of Shell objects
+      type(gauxc_shell_type), intent(in) :: shells(*)
+      !> @param nshells Number of shells in the array
+      integer(c_size_t), value :: nshells
+      !> @return Pointer to the newly created basis set object
+      type(gauxc_basisset_type) :: basis
+    end function gauxc_basisset_new_from_shells
+  end interface
```
kjh-giesbertz: Interface the additional basis set routines also from fortran
```suggestion

    integer(c_int32_t) function gauxc_basisset_nshells(status, basis) bind(c)
      import :: c_int32_t, gauxc_status_type, gauxc_basisset_type
      implicit none
      !> @param status Status of the operation
      type(gauxc_status_type), intent(out) :: status
      type(gauxc_basisset_type), value :: basis
    end function

    integer(c_int32_t) function gauxc_basisset_ngtos(status, basis) bind(c)
      import :: c_int32_t, gauxc_status_type, gauxc_basisset_type
      implicit none
      !> @param status Status of the operation
      type(gauxc_status_type), intent(out) :: status
      type(gauxc_basisset_type), value :: basis
    end function
  end interface
```

src/c-api/c_runtime_environment.cxx
```diff
@@ -0,0 +1,112 @@
...
+    env.ptr = new RuntimeEnvironment(GAUXC_MPI_CODE(comm));
+    env.hdr = GauXCHeader{GauXC_Type_RuntimeEnvironment};
+    status->code = 0;
+  } catch (std::exception& e) {
+    status->code = 1;
+    status->message = detail::strdup(e.what());
+  }
+  return env;
+}
+
```
kjh-giesbertz, line 41: Add a variant which additionally transforms a fortran MPI communicator to the c-version
```suggestion

GauXCRuntimeEnvironment gauxc_runtime_environment_new_fortran(
  GauXCStatus* status
  GAUXC_MPI_CODE(, MPI_Fint f_comm)
) {
  MPI_Comm c_comm;
  c_comm = MPI_Comm_f2c(f_comm);
  return gauxc_runtime_environment_new(status GAUXC_MPI_CODE(, c_comm));
}

```

awvwgk:
Current version fails with OpenMPI due to missing f2c API in mpi_f08 module.

wavefunction91:
## 🦆 Rubber Duck Review

I ran a multi-agent rubber-duck review over the Fortran API, deploying five parallel reviewers each focused on a different slice (build/CMake, MPI&ISO_C_BINDING, lifecycle/small wrappers, the big `xc_integrator.f90`, and functional+tests). The two most consequential findings (`svwn3`/`svwn5` enum bug and the integrator-wrapper contiguity claim) I verified by hand.

Findings are ranked by severity; file:line citations throughout.

### Critical — silent wrong results or memory corruption

1. **`svwn3` / `svwn5` enum aliases map to the wrong enum values** — `src/fortran-api/functional.f90:351-352`
   ```fortran
   integer(c_int) :: svwn3 = gauxc_functional_vwn3   ! ← should be _svwn3
   integer(c_int) :: svwn5 = gauxc_functional_vwn5   ! ← should be _svwn5
   ```
   SVWN = Slater exchange + VWN correlation; VWN = correlation-only. A user who writes `gauxc_functional%svwn3` silently gets a *correlation-only* functional and a physically wrong total energy. The raw `gauxc_functional_svwn3` / `_svwn5` enumerators are fine; only these two struct aliases are wrong. No test exercises them.

2. **`strlen` interface declared to return `c_int`, but C `strlen` returns `size_t`** — `src/fortran-api/status.f90:34-35` — 4-byte read of an 8-byte register on LP64. Works by luck when the upper bits happen to be zero. Fix: `integer(c_size_t) :: len` and matching `slen`.

3. **`MPI_Fint` ↔ `integer(c_int)` mismatch** — `src/fortran-api/runtime_environment.F90:61,118,135` and `src/fortran-api/f_runtime_environment.cxx:20,28,36`. `MPI_Fint` is implementation-defined. Works on every common Linux MPI today, but Cray MPICH and MPICH `--enable-large-tests` can make it 64-bit, in which case the Fortran handle is silently truncated before `MPI_Comm_f2c`. Either expose the real `MPI_Fint` kind via a CMake-probed preprocessor define, or add `static_assert(sizeof(MPI_Fint) == sizeof(int), …)` in `f_runtime_environment.cxx` to make the assumption explicit.

4. **Status message memory leak — no Fortran-accessible `gauxc_status_delete`** — `src/fortran-api/status.f90` — Every error path on the C side allocates `status->message`; the Fortran `gauxc_status` module never wraps `gauxc_status_delete`, so messages accumulate over the lifetime of the program. Expose the delete and document that callers must call it (or have the wrappers call it after the user reads the message).

5. **`GAUXC_HAS_MPI_F08` set TRUE unconditionally when MPI is on** — `src/fortran-api/CMakeLists.txt:41-43` — No probe that the installed MPI actually provides the `mpi_f08` module. MPICH <3.1, OpenMPI 2.x and several vendor MPIs do not. Build fails deep in the compiler with no CMake-level explanation. Use `check_fortran_source_compiles("program p; use mpi_f08; end program" HAVE_MPI_F08)` and set the flag from that result.

6. **Test driver guards `use mpi_f08` with the wrong macro** — `tests/fortran_main.F90:7-8` uses `#ifdef GAUXC_HAS_MPI`, but the library everywhere else uses the nested `#ifdef GAUXC_HAS_MPI_F08`. Once #5 is fixed, the test will fail to compile on legacy MPI installs.

7. **Assumed-shape arrays passed to `bind(c)` assumed-size dummies without `contiguous`** — `src/fortran-api/molecule.f90:111-120` (`atoms(:)`) and `src/fortran-api/basisset.f90` (`shells(:)`). If a user passes a strided slice (e.g. `atoms(1::2)`), the C side reads garbage. Add the `contiguous` attribute. (The xc_integrator wrappers correctly use `contiguous` already — confirmed at `src/fortran-api/xc_integrator.f90:606-612`.)

### High

| # | Issue | Location |
|---|---|---|
| 8 | `enable_language(Fortran)` in installed `gauxc-config.cmake` fires for **every** consumer, including pure C++ projects — hard fails on Fortran-less CI/HPC images | `cmake/gauxc-config.cmake.in:32` |
| 9 | `.mod` files installed flat to `include/gauxc/modules/` — not portable across compilers or even gfortran major versions; silently breaks downstream | `src/fortran-api/CMakeLists.txt:54-58` |
| 10 | `hdf5_read.cxx` and `hdf5_read.f90` (and `_write`) share leaf names in same target — collide in VS / Xcode generators that key off the leaf | `src/external/CMakeLists.txt:35,40` |
| 11 | Test main runs all assertions on every MPI rank → duplicated output and rank-dependent exit codes | `tests/fortran_main.F90:32` |
| 12 | Multiple tests leak GauXC objects on early-exit error paths, masking real library leaks under Valgrind/ASan | `tests/fortran_api_test.f90:170-207, 272-286, 313-397` |
| 13 | `use mpi, only: MPI_Comm` (non-f08 branch) — `MPI_Comm` is not guaranteed exported from MPICH's `mpi` module; only needed in the f08 branch anyway | `src/fortran-api/runtime_environment.F90:22` |
| 14 | `%mpi_val` is MPI-4 (lowercase is an OpenMPI extension); not standard in MPI-3 f08 | `src/fortran-api/runtime_environment.F90:171,196,210,241,255,288` |
| 15 | `gauxc_runtime_environment_new_f` symbol arity changes by build flag without renaming → mixed-MPI binaries can link but crash at runtime | `src/fortran-api/f_runtime_environment.cxx:18-23` |
| 16 | `gauxc_delete` is a *per-module* generic; user must `use` every relevant module to see all specifics — undocumented footgun | (multiple modules) |
| 17 | `hdr.type` not cleared after `gauxc_*_delete`, so zombie handles still report their original tag | (every handle type) |

### Medium

`cmake_dependent_option` silently demotes `GAUXC_ENABLE_FORTRAN=ON` when C is off (no warning); `GAUXC_HAS_FORTRAN` set in two places that can drift; `include/gauxc/gauxc_config.f.in` lacks an include guard; `cmake/gauxc-test-drive.cmake` falls through to a nonexistent `test-drive-lib` target on miss; no `static_assert` on `GauXCRuntimeEnvironment` size to catch `GAUXC_HAS_DEVICE` skew; HDF5 wrappers declare `status` as `intent(out)` inconsistent with the rest of the API (`src/external/hdf5_read.f90:33,67`, `hdf5_write.f90:33,67`); `use iso_c_binding` missing `intrinsic` qualifier in `functional.f90:13`, `molecular_weights.f90:13`, `hdf5_read.f90:13`, `hdf5_write.f90:13`; `MPI_Initialized` imported but never called; no numerical integration test — pipeline test only checks header tags; functional-enum-struct values never verified by tests, so finding #1 is invisible.

### Low

`value` attribute on user-facing Fortran wrappers (unusual style); dead `MAKE_DIRECTORY` at configure time; `gauxc_type_*_factory` enums publicly exported but otherwise unused; generic `gauxc_delete` never exercised by a test.

### What the reviewers liked

- **`src/fortran-api/xc_integrator.f90` (860 LOC) came back essentially clean.** All 14 `bind(c)` interfaces match their C counterparts; `c_int` vs `c_int64_t` is consistent throughout; null-termination handled via `transfer`+`c_null_char`; GKS 4-component splitting correct; FXC separates `P`/`T` with independent leading dims; `contiguous` used on every assumed-shape wrapper.
- 79-functional enum ordering matches the C side exactly (only the two struct aliases above are wrong).
- `gauxc_atom_type` struct layout matches C with no padding surprises (`int64_t Z` then three doubles, natural alignment).
- C side reliably sets `ptr = nullptr` on factory error paths, so post-failure `gauxc_delete` is safe.
- `MPI_Comm_f2c` placed at the correct C/Fortran boundary — textbook design.
- `cmake/gauxc-test-drive.cmake` pins to a real version (`v0.5.0`) and tries `CONFIG`/`PkgConfig` before fetch.
- `.F90` vs `.f90` naming correctly applied throughout.

### Bottom line

Most of the codebase is solid, careful Fortran. The integrator wrapper — the file most exposed to silent numerical bugs — is exemplary. The defects cluster in three places that genuinely benefit from a careful Fortran review: **(a) the enum-struct copy-paste bug**, **(b) MPI portability assumptions** (`MPI_Fint`, `mpi_f08` always present, `MPI_Comm` export, `%mpi_val`), and **(c) install-time portability** (`.mod` files, downstream `enable_language(Fortran)`). I'd treat findings 1, 2, 6, 7 as merge-blockers; 3, 4, 5 as required follow-ups; and the High items as issues to file before close.

<sub>Generated by a multi-agent rubber-duck review (5 parallel code-review agents, one per scope) via GitHub Copilot CLI.</sub>

PR #175, opened 2026-01-29
Fetch dependencies as tarballs and verify checksums

awvwgk:
Download dependencies via FetchContent as tarball rather than git repo, verify tarballs based on their checksum.


awvwgk:
Not needed.

PR #176, opened 2026-02-06
Add sanity check for number of primitives

susilehtola:
There's no check in GauXC on the maximum allowed number of primitives. I think any faulty code might error out even earlier due to overflows in the passed-in arrays, but this check will at least crash the code after the memory violation has occurred instead of producing faulty results.

### DIFF
--- a/include/gauxc/shell.hpp
+++ b/include/gauxc/shell.hpp
@@ -138,6 +138,9 @@ public:
     alpha_( alpha ), coeff_( coeff ), O_( O ),
     nprim_( nprim.get() ), l_( l.get() ), pure_( pure.get() ) {
 
+    if(nprim_ > detail::shell_nprim_max)
+      throw std::out_of_range("Shell has too many primitives!\n");
+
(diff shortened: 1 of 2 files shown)


include/gauxc/shell.hpp
```diff
@@ -138,6 +138,9 @@ class alignas(256) Shell {
     alpha_( alpha ), coeff_( coeff ), O_( O ),
     nprim_( nprim.get() ), l_( l.get() ), pure_( pure.get() ) {
 
+    if(nprim_ > detail::shell_nprim_max)
+      throw std::out_of_range("Shell has too many primitives!\n");
```
wavefunction91: Might be good to add the actual values (max val + actual val) here to aid users in finding the problem

wavefunction91 (changes requested):
Good call. Minor comment

wavefunction91 (approved):

awvwgk (approved):
Looks good to me.

PR #180, opened 2026-02-12
Update tests to use catch2 v3

### LINKED ISSUE #178: Cannot find catch2 dependency
When building and installing ExchCXX and IntegratorXX and using them as dependency GauXC cannot find catch2 for building the tests. It seems that the GauXC test relies on either of these dependencies to make catch2 available.

awvwgk:
Closes #178 
Closes #179


wavefunction91 (approved):
I'm generally fine with this, but you will need to update with `master`

awvwgk:
David @wavefunction91, can we merge this?

PR #186, opened 2026-03-16
Fix GPU MPI coords device pointer bug

tvogels:
In the multi-rank MPI path, `prepare_onedft_features` was called with CPU `TensorOptions` but passed `device_data_ptr->coords_device_data()`, which is a GPU pointer. This is undefined behavior — a CPU tensor wrapping device memory.

This fix builds `host_coords` from the `Molecule` object (already on the host) instead of using the device pointer.

We probably missed this because the multi-rank case is not as well-covered by tests as the single rank path?


awvwgk (approved):
Good catch, thank you.

PR #190, opened 2026-03-31
Add raw_weights field to XCTask for pre-partition quadrature weights

tvogels:
Adds a `raw_weights` vector to `XCTask` that preserves the original quadrature weights before molecular partition scaling. With this change, neural network functionals can make use of atomic grids for internal integration.

The `merge` and `merge_with` helpers are updated to propagate `raw_weights` when present.


wavefunction91:
@tvogels Thanks. What's the use case for this? That's a lot of storage for large problems. Could be O(GB).

PR #193, opened 2026-04-13
Fix Fortran compile error: Remove unnecessary parameter from binding

franzpoeschel:
#### Note that this is NOT a PR to the master branch

The Fortran bindings on the Skala branch currently fail compiling:

```
/home/user/git-repos/cp2k/venv/GauXC/src/fortran-api/molecule.f90:111:62:

  111 |   function gauxc_molecule_new_from_atoms(status, atoms, natoms) result(molecule)
      |                                                              1
Error: Symbol 'natoms' at (1) has no IMPLICIT type; did you mean 'atoms'?
```

`natoms` is not used, the atom count is determined from the size of the `atoms` array.

### DIFF
--- a/src/fortran-api/molecule.f90
+++ b/src/fortran-api/molecule.f90
@@ -108,7 +108,7 @@ module gauxc_molecule
 contains
 
   !> @brief Create a new Molecule instance from an array of Atoms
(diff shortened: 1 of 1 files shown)


awvwgk:
Thanks, I included this in the Skala branch and also updated the tarball in the skala repo.

PR #194, opened 2026-04-15
Add PySCF grid levels to AtomicGridSizeDefault values

awvwgk:
For reproducing PySCF grids in GauXC it would great to have them available as part of the `AtomicGridSizeDefault` enumerator.


wavefunction91:
Thanks, @awvwgk. I'm in principle supportive of this, but the size is likely the least of the concerns here - can we get an example of us actually reproducing the grids (e.g. w/pruning)? I'm fine having the path, but I have some severe reservations about how they prune grids. It would be good to have that in the open so we can do AB...Z comparisons.

awvwgk:
Can do, but I mainly care about the grid sizes here. Does GauXC already implement the same pruning scheme as used in PySCF or does this need to be implemented separately first?

awvwgk:
See https://github.com/wavefunction91/GauXC/pull/198 for an implementation of the pruning schemes.

Tests in this PR comprehensively check the grid size definition for all elements against the values obtained by the `pyscf.dft.gen_grid` module for the angular and radial sizes. In #198 I am testing against selected cases for each of the PySCF pruning schemes, not sure whether making it comprehensive is useful (_num elements_ x _num grid sizes_ x _pruning schemes_ = ~5k).

wavefunction91 (approved):

PR #199, opened 2026-04-27
Add normalize parameter to gauxc_basisset_new_from_shells

franzpoeschel:
This is the C function bound to Fortran here. Leaving the normalize param unspecified here seems like undefined behavior.

```c
/**
 * @brief Create a new BasisSet instance from arrays of shells.
 * @param status Status object to capture any errors.
 * @param shells Pointer to an array of GauXCShell.
 * @param nshells Number of shells in the array.
 * @param normalize Whether to normalize the basis functions.
 * @return Handle to the created BasisSet.
 */
extern GauXCBasisSet gauxc_basisset_new_from_shells(
   GauXCStatus* status,
   GauXCShell* shells,
   size_t nshells,
   bool normalize
);
```

### DIFF
--- a/src/fortran-api/basisset.f90
+++ b/src/fortran-api/basisset.f90
@@ -10,7 +10,7 @@
 
 !> @brief Module defining basis set functionality for GauXC
 module gauxc_basisset
-  use iso_c_binding, only : c_ptr, c_null_ptr, c_size_t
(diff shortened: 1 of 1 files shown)


awvwgk:
Thanks for the feedback, implemented in https://github.com/wavefunction91/GauXC/pull/174 and added to the skala branch.

Issue #79, opened 2023-10-24
Density Inputs / Potential Outputs Should be Consistent and Flexible for RKS and UKS


wavefunction91:
As of #78, UKS adopts a [canonical Pauli definition ](http://wavefunction91.github.io/_assets/spinor_dft.pdf) of the density input / VXC output. RKS currently takes $D_\alpha$, which is not consistent with the Pauli definition. Looking forward to GKS, the Pauli implementation is superior as all input densities are Hermitian (as is not the case in the spin-separated convention),  while one can go either way for RKS/UKS. We should also support the latter to allow for simple integration with spin-separated codes (which are the norm).

1. RKS needs to (optionally) accept $D_s$ as input - this could be resolved by a strong type template parameter toggling the expected input. Luckily, the potential remains the same for either spin-separated or Pauli definitions of the density / potential. 
2. UKS needs to (optionally) accept $D_\alpha$ / $D_\beta$ as input and return $V_\alpha$ / $V_\beta$ as output. This _**does**_ change the expected output by $\pm$ factors when this option would be enabled

wavefunction91:
Ping @evaleev, @ajaypanyala, @loriab, @susilehtola, @awvwgk 
CC @DCM-Uni-Paderborn

Any opposition to changing the convention for RKS to take `N` as opposed to `\alpha`? Ref #208

ajaypanyala:
@wavefunction91 Sounds good. No issues on our end. Thanks for checking!

wavefunction91:
Ping @mikovtun for CQ

mikovtun:
@wavefunction91 No issues from CQ either, thanks for the ping

DCM-Uni-Paderborn:
Following the stakeholder feedback above and the downstream CP2K fix in cp2k/cp2k#5740, I would like to settle the remaining library-level design before writing more code. Would GauXC prefer: (A) making the spin-summed density N = P_alpha + P_beta the single RKS input convention across EXC, VXC, EXC gradients, and FXC, without a runtime flag, together with updated tests, documentation, and a migration note; or (B) retaining the current one-spin RKS convention and requiring callers to convert locally? If (A) is preferred, I will supersede #208 with a new, narrowly scoped implementation against current master. If (B) is preferred, I will close #208 because CP2K now performs the conversion safely downstream. The high-angular-momentum EXC-gradient discrepancy is independent of this convention question and will be tracked separately with the standalone HDF5 reproducer.

PR #207, opened 2026-05-24
Expose spin-summed RKS setting through C/Fortran APIs

DCM-Uni-Paderborn:
Dear David,

this is the follow-up to #208 for the `skala` branch. It assumes that the core C++ setting and host/device/shell-batched handling from #208 are accepted first, and only exposes that setting through the external APIs needed by downstream callers such as CP2K.

Changes in this PR are now limited to:

- C API `GauXCKSSettings`
- C API RKS entry points with explicit settings for EXC, EXC/VXC, EXC gradients, and FXC contraction
- Fortran API optional RKS settings wrappers
- a focused C API regression check for spin-summed closed-shell RKS density input

The core `IntegratorSettingsKS` extension, host/device/shell-batched RKS scaling, standalone-driver keyword, and C++ integrator regression are intentionally left to #208 and are no longer duplicated here.

This remains a draft until #208 is merged and the `skala` branch contains the agreed core convention.

Validation:

- checked `git diff --check`
- reduced the diff against `skala` to the C API, Fortran API, and C API test files only

Greetings,
Thomas

### DIFF
--- a/include/gauxc/c/xc_integrator.h
+++ b/include/gauxc/c/xc_integrator.h
@@ -37,6 +37,14 @@ typedef struct GauXCIntegrator {
 } GauXCIntegrator;
 
 /**
+ * @brief Settings for Kohn-Sham XC integrations.
+ */
+typedef struct GauXCKSSettings {
+  double gks_dtol; ///< Density tolerance used by GKS integrations.
+  bool rks_density_matrix_is_spin_summed; ///< True if an RKS matrix contains the spin-summed density.
+} GauXCKSSettings;
+
+/**
  * @brief Delete an XCIntegrator instance.
  * @param status Status object to capture any errors.
  * @param integrator Handle to the XCIntegrator to delete.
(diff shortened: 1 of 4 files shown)


awvwgk (changes requested):
Can we separate the changes to the Skala implementation and core GauXC? The extension of the settings struct is a potential API change for the core library that is unrelated to Skala and is better discussed independently of the changes necessary to expose the new setting to the Fortran API for Skala.

DCM-Uni-Paderborn:
Dear Sebastian,

Yes, that makes sense. I split off the core GauXC part into #208 against `master`. That PR only adds the C++ setting and the corresponding host/device/shell-batched RKS handling, plus a focused C++ regression check; it does not include Skala, OneDFT, C API, or Fortran API exposure.

I will keep this `skala`-branch PR as draft and rework/rebase it after the core convention is settled, so the follow-up here only exposes the agreed setting through the Skala/Fortran-facing API needed by CP2K.

Thanks for pointing this out.

DCM-Uni-Paderborn:
Closing this API-layer PR because it is coupled to the runtime convention flag proposed in #208, whereas the maintainer discussion favors either a single consistent RKS convention or caller-side conversion. CP2K's concrete normalization and call-order bug has meanwhile been fixed downstream in cp2k/cp2k#5740 by passing private one-spin working copies while leaving the spin-summed CP2K density matrix unchanged. The broader GauXC convention decision remains in #79/#208. The separate high-angular-momentum EXC-gradient discrepancy is not addressed by this PR and will be handled independently.

PR #213, opened 2026-06-08
Fix CUTLASS local work driver setup

awvwgk:
See #206

### DIFF
--- a/cmake/gauxc-exchcxx.cmake
+++ b/cmake/gauxc-exchcxx.cmake
@@ -1,5 +1,5 @@
 find_package( ExchCXX QUIET )
-if( NOT ${ExchCXX_FOUND} )
+if( NOT ExchCXX_FOUND )
 
   include( gauxc-dep-versions )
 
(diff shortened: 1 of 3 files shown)


cmake/gauxc-exchcxx.cmake
```diff
@@ -19,6 +19,10 @@ if( NOT ${ExchCXX_FOUND} )
 
   FetchContent_MakeAvailable( exchcxx )
 
+  if( CMAKE_POSITION_INDEPENDENT_CODE AND TARGET exchcxx )
```
wavefunction91, line 22: I'm fine with this in principle, but can you point to the point where this caused an error for the current bug fix? This applies to the other instance as well.

src/xc_integrator/local_work_driver/factory.cxx
```diff
@@ -11,18 +11,31 @@
...
 #include "host/reference_local_host_work_driver.hpp"
+#include <algorithm>
+#include <cctype>
 #ifdef GAUXC_HAS_DEVICE
 #include "device/cuda/cuda_aos_scheme1.hpp"
 #include "device/hip/hip_aos_scheme1.hpp"
 #endif
 
 namespace GauXC {
+namespace {
```
wavefunction91, line 22: Please done use anonymous namespaces. "detail" would be appropriate here.

src/xc_integrator/local_work_driver/factory.cxx
```diff
@@ -11,18 +11,31 @@
...
+
+void trim_and_uppercase(std::string& name) {
```
wavefunction91, line 24: Since this is only used once, I don't see the value in adding the helper function.

wavefunction91 (changes requested):

PR #217, opened 2026-07-16
Fix mpi bug

JensWehner:
Fix for https://github.com/cp2k/cp2k/issues/5537

as described there, the problem was using the global communicator, because cp2k actually passes a subgroup in and then indices derived from the global one are wrong. Exactly what they described in the issue.

Added a test and compiled cp2k with it and then the error seems to be gone.


tvogels (approved):
Looks good, thanks!

awvwgk (approved):
Thanks, looks good to me.

PR #223, opened 2026-08-29
Align CUDA sources with C++17

JensWehner:
GauXC’s CUDA sources are not strictly C++14-compatible because they already use C++17 features such as if constexpr. However, the build configuration declares cuda_std_14, while several kernels still use the register keyword, which was removed in C++17.

This PR aligns the declared CUDA standard with the code by:

Changing cuda_std_14 to cuda_std_17.
Removing obsolete register qualifiers that Clang diagnoses in C++17 mode.


awvwgk (approved):
Thanks for making this consistent across the codebase.

